PR #7, opened 2018-09-10
[MJAVADOC-537] Explicitly the batchMode to true

joshiste:
To prevent a warning about a missing input stream we explicitly set
batchMode when invoking other maven commands

fixes MJAVADOC-537


jira-importer:
Resolve #839

PR #10, opened 2018-11-13
[MJAVADOC-544] - Changed behaviour of Javadoc for temporary files encoding (options, argfile, ...)

michael-st:
mvn -Prun-its verify was run with JDK9 on windows without errors

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -108,6 +108,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
@@ -4374,9 +4375,15 @@ public abstract class AbstractJavadocMojo
         options.append( StringUtils.join( arguments.iterator(),
                                           SystemUtils.LINE_SEPARATOR ) );
 
+        /* default to platform encoding */
+        String outputFileEncoding = null;
+        if ( JAVA_VERSION.isAtLeast( "9" ) )
+        {
+            outputFileEncoding = StandardCharsets.UTF_8.name();
+        }
         try
         {
-            FileUtils.fileWrite( optionsFile.getAbsolutePath(), null /* platform encoding */, options.toString() );
+            FileUtils.fileWrite( optionsFile.getAbsolutePath(), outputFileEncoding, options.toString() );
         }
         catch ( IOException e )
         {
--- a/pom.xml
+++ b/pom.xml
@@ -116,6 +116,9 @@ under the License.
     <contributor>
       <name>Kevin Risden</name>
     </contributor>
+    <contributor>
+      <name>Michael Stumpf</name>
+    </contributor>
   </contributors>
 
   <dependencies>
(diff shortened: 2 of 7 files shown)


michael-st:
Due to changes to readFile in b78042ec conflicts were introduced.

I rebased the PR and fixed some style-issues in the tests.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -4368,9 +4378,15 @@ private void addCommandLineOptions( Commandline cmd, List<String> arguments, Fil
         options.append( StringUtils.join( arguments.iterator(),
                                           SystemUtils.LINE_SEPARATOR ) );
 
+        /* default to platform encoding */
+        String encoding = null;
+        if ( JAVA_VERSION.compareTo( SINCE_JAVADOC_9 ) >= 0 )
```
rfscholte: Please switch to `isAtLeast()` or `isBefore()`, which are better readable.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -4414,9 +4430,15 @@ private void addCommandLineArgFile( Commandline cmd, File javadocOutputDirectory
             cmd.createArg().setValue( "@" + FILES_FILE_NAME );
         }
 
+        /* default to platform encoding */
+        String encoding = null;
+        if ( JAVA_VERSION.compareTo( SINCE_JAVADOC_9 ) >= 0 )
```
rfscholte: Please switch to `isAtLeast()` or `isBefore()`, which are better readable.

rfscholte: Don't forget to squash your commits

michael-st: I addressed your points and squashed the commits.

Ran mvn -Prun-its verify with JDK8 and JDK9 on Windows successfully.

rfscholte:
https://builds.apache.org/view/M-R/view/Maven/job/maven-box/job/maven-javadoc-plugin/job/MJAVADOC-544/ we're not there yet.

michael-st:
too bad...

The error message occurs when compiling the file "/src/test/resources/unit/argfileumlautencoding-test/argfileumlautencoding/test/AppÃ¤Ã¶Ã¼Ã.java".
The compiler complains about a name mismatch between class "Appäöüß" and the filename "AppÃ¤Ã¶Ã¼Ã.java".
This is a charset-problem, which occurs only in the linux environment.

I tried to reproduce the problem:

The build passes when I execute the maven build in a docker container (image: maven:3-jdk-9-slim) in which the LANG is set to C.UTF-8.
If I change the LANG-environment variable in the container ("export LANG=en_US.ISO-8859-1") - the build fails.

```
FROM maven:3-jdk-9-slim
RUN echo $LANG
> C.UTF-8
RUN apt update && apt install -y git && echo dirty no cleanup of apt
RUN git clone https://github.com/michael-st/maven-javadoc-plugin.git
RUN cd maven-javadoc-plugin
RUN git checkout MJAVADOC-544
RUN mvn -Dtest=org.apache.maven.plugins.javadoc.JavadocReportTest#testArgfileUmlautEncoding clean test
> build passes
RUN export LANG=en_US.ISO-8859-1
RUN echo $LANG
> en_US.ISO-8859-1
RUN mvn -Dtest=org.apache.maven.plugins.javadoc.JavadocReportTest#testArgfileUmlautEncoding clean test
> build fails
```

I tried to convert the file itself and the corresponding .xml-file to ISO-8859-1, but that did not go well either.

At the moment I'm out of ideas and I don't think that I can find a working solution for this (besides setting the LANG to ...UTF-8 in the build environment).

My initial problem was only related to the options file, which would be fixed with this pull request.
For this all tests should pass (hopefully).

So, for this issue, I have removed the special handling for argfile, which would only be needed if someone is using special characters in java class names.
I also modified the commit message to reflect this change.

Is this course of action okay?

PR #14, opened 2018-12-24
Exclude non-exported packages when Java Modules are present.

cowwoc:


src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -2514,7 +2523,7 @@ protected boolean canGenerateReport( List<String> files )
      * Method to get the excluded source files from the javadoc and create the argument string
      * that will be included in the javadoc commandline execution.
      *
-     * @param sourcePaths the collection of paths to the source files
+     * @param sourceFolders the collection of paths to the source files
```
cowwoc: Minor bugfix: The Javadoc parameter name did not match the method parameter name.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -2015,6 +2013,17 @@ protected void executeReport( Locale unusedLocale )
         }
         setFJavadocVersion( new File( jExecutable ) );
 
+        List<String> packageNames;
```
cowwoc: I had to move these lines under `setFJavaVersion()` because `javadocRuntimeVersion` is not set before this point.

rfscholte:
Can you add your testcase to `src/it` to confirm this fixes the assue and to prevent regression in the future.

rfscholte:
One of the most important instructions on https://github.com/apache/maven-javadoc-plugin is:
> Run all the tests with `mvn -Prun-its verify` to assure nothing else was accidentally broken.
What you do next is put your testcase under `src/it/MJAVADOC-XXX_some-keywords` and let the maven-invoker-plugin do the rest. https://maven.apache.org/plugins/maven-invoker-plugin/usage.html describes its usage (don't forget to replace the plugin version with `@project.version@`.)

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -4342,6 +4351,102 @@ private void copyAdditionalJavadocResources( File anOutputDirectory )
...
+    private List<String> getPackageNamesRespectingJavaModules( Map<String, Collection<String>> allSourcePaths )
+            throws MavenReportException
+    {
+        List<String> returnList = new ArrayList<>();
+
+        if ( !StringUtils.isEmpty( sourcepath ) )
+        {
+            return returnList;
+        }
+        LocationManager locationManager = new LocationManager();
```
rfscholte: I hope we can refactor the code, so the locationManager is used only once per goal-call.

cowwoc: This method only gets invoked once per goal-call. Anyway, feel free to refactor this code to your liking. I don't mind :)

rfscholte: locationManager is used at least twice now, which is pretty expensive. I'll refactor it

rfscholte:
Merged into Apache Git

PR #15, opened 2018-12-24
Don't fail if module source path already exists.

cowwoc:
Improved error message if IOException is thrown (added chained exception).


src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -4754,11 +4754,11 @@ private void addJavadocOptions( File javadocOutputDirectory,
                             addArgIfNotEmpty( arguments, "--patch-module", moduleName + '='
                                 + JavadocUtil.quotedPathArgument( getSourcePath( projectSourcepaths.getValue() ) ) );
                             
-                            Files.createDirectory( moduleSourceDir.resolve( moduleName ) );
+                            Files.createDirectories( moduleSourceDir.resolve( moduleName ) );
```
eolivelli: The semantics of this method are different from the ones of createDirectory.
If you only want to create the last path you should use
if !Files.isDirectory...createDirectory

rfscholte: That's probably better: `moduleSourceDir` should already exist and `moduleName` should reflect 1 directory.

cowwoc:
@eolivelli @rfscholte I'm new to integration tests in this plugin. I tried adding an integration test in https://github.com/apache/maven-javadoc-plugin/pull/15/commits/a83d81d85069a806ad3467a0c092b60476829f4a that should fail, but it keeps on passing. Can you please help me figure out what I'm doing wrong? Thank you.

rfscholte:
In this case an integration test is a little bit too much. The code is simple, touched often enough by other tests and a unittest takes quite some time to run.

cowwoc:
@rfscholte Fair enough, but I still need to understand how to write proper integration tests because https://github.com/apache/maven-javadoc-plugin/pull/14 will need one. Can you please help me understand why this one is not failing as expected?

cowwoc:
@eolivelli I am trying to invoke `javadoc:aggregate` twice in a row. I tried https://github.com/apache/maven-javadoc-plugin/pull/15/commits/8960026ed32ef3a666f5c44000a40a49aed66577 but it seems to invoke `javadoc:aggregate` then run other goals then run `javadoc:aggregate` a second time. The intermediate goals are changing the state of the project and preventing the second `javadoc:aggregate` invocation from failing.

I just committed https://github.com/apache/maven-javadoc-plugin/pull/15/commits/4beb5c9f776e8e506bc4ee154228022227dcfb4f to remove the integration test per Robert's feedback, but I'd really like to know how I was supposed to get the integration test working as expected.

As an aside, I believe this branch should now be ready to merge.

cowwoc:
@eolivelli Nevermind, I figured it out. There were numerous things wrong with that commit. Namely:

* The top-level pom was missing a `<modules>` section to load the sub-module.
* The child module referenced the wrong parent artifact.
* The child module needed a compiler plugin configured to build under Java 9+.

Anyway, now that this is out of the way the only thing remaining is to merge this PR into `master`.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
cowwoc: Done. I've reworked the fix in terms of `if !Files.isDirectory...createDirectory`.

cowwoc:
@khmarbaise Fixed. Thank you!

rfscholte:
Merged in Apache Git

PR #19, opened 2019-01-26
Avoid converting separator characters into newline.

cowwoc:
Updated testCustomConfiguration() because JDK 11.0.2 behaves the same as JDK 12.


jira-importer:
Resolve #855

PR #21, opened 2019-02-13
[MJAVADOC-578] Groups parameter is not compatibile with Surefire

eolivelli:
See https://issues.apache.org/jira/browse/MJAVADOC-578 for reference

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -1043,7 +1043,7 @@ public abstract class AbstractJavadocMojo
      * See <a href="./apidocs/org/apache/maven/plugin/javadoc/options/Group.html">Javadoc</a>.
      * <br/>
      */
-    @Parameter( property = "groups" )
+    @Parameter
     private Group[] groups;
 
     /**


src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -1043,7 +1043,7 @@
      * See <a href="./apidocs/org/apache/maven/plugin/javadoc/options/Group.html">Javadoc</a>.
      * <br/>
      */
-    @Parameter( property = "groups" )
+    @Parameter()
```
rfscholte: And no need for parentheses anymore

eolivelli: @rfscholte 
sure, thank you, removed

eolivelli:
@rfscholte I think we don't need an integration test for this change

eolivelli:
Link to CI results
https://builds.apache.org/job/maven-box/job/maven-javadoc-plugin/job/MJAVADOC-578

rfscholte:
> I think we don't need an integration test for this change

I agree. Removing an impossible user property to use should have no effect.

eolivelli:
Good.
I will merge as soon as I have a "blue" build on Jenkins

PR #25, opened 2019-04-13
[MJAVADOC-591] consider release for supportModulePath 

rovarga:
Since we are setting javadoc -release to whatever is set in maven.compiler.release, we need to check if we are not using it to downgrade javadoc before attempting to add modulePath-related options.

Signed-off-by: Robert Varga <robert.varga@pantheon.tech>

       pulling in other changes.
       commit message.

### DIFF
--- a/src/it/projects/MJAVADOC-591/pom.xml
+++ b/src/it/projects/MJAVADOC-591/pom.xml
@@ -1,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one
+  or more contributor license agreements.  See the NOTICE file
+  distributed with this work for additional information
+  regarding copyright ownership.  The ASF licenses this file
+  to you under the Apache License, Version 2.0 (the
+  "License"); you may not use this file except in compliance
+  with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing,
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied.  See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>testcase</groupId>
+  <artifactId>javadoc-fails-with-release-and-not-source</artifactId>
+  <version>1.0-SNAPSHOT</version>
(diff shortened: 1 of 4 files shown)


src/it/projects/MJAVADOC-591/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+  <groupId>testcase</groupId>
+  <artifactId>javadoc-fails-with-release-and-not-source</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <packaging>bundle</packaging>
+
+  <url>https://issues.apache.org/jira/browse/MJAVADOC-591</url>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.compiler.release>8</maven.compiler.release>
```
eolivelli: One question: your new line of code works with java >= 9 but this test uses 8.
Is this test case covering your change?

src/it/projects/MJAVADOC-591/src/main/java/com/example/Foo.java
```diff
@@ -0,0 +1,31 @@
+package com.example.foo;
```
rfscholte: change package to `com.example`, otherwise the integration-testwill fail.

src/it/projects/MJAVADOC-591/pom.xml
rovarga: I actually have some trouble getting ITs running locally and I know next to nothing about their setup -- but this IT should run with JDK9+ (via invoker.properties, if I understand correctly). The maven.compiler.release setting is correct, as it mirrors the scenario where I hit the issue -- the combination of https://github.com/opendaylight/odlparent/blob/master/odlparent/pom.xml#L1878 and https://github.com/opendaylight/yangtools/blob/master/common/concepts/pom.xml#L42 build with JDK11.

src/it/projects/MJAVADOC-591/src/main/java/com/example/Foo.java
rovarga: Done, thanks.

src/it/projects/MJAVADOC-591/pom.xml
rfscholte: You can only use the `release`-flag when running with Java 9+. So that's correct. Saying release 8 means that your code is Java 8 compatible, even when running with a newer version of Java. (this was not guaranteed with Java 8 and before + `source`, hence animal-sniffer)

eolivelli: Got it
Thanks

And the fact the this reproduces your case to me is the most important thing.
We are sure we are fixing your problem

rovarga: The test case triggers correctly, without the fix running with JDK12 I get:

[INFO] Building: MJAVADOC-591/pom.xml
[INFO]           MJAVADOC-591/pom.xml ............................. FAILED (5.2 s)
[INFO]   The build exited with code 1. See /home/nite/prj/maven-javadoc-plugin/target/it/MJAVADOC-591/build.log for details.

with the fix applied the test passes.

rfscholte:
Fixed with https://github.com/apache/maven-javadoc-plugin/commit/56979750c1752c811da28ff058aebc83b1fd018b

PR #34, opened 2019-11-03
[MJAVADOC-628] - add javadoc.io link pattern when <detectLink> option is …

maxcellent:
…set to true

       pulling in other changes.
       commit message.

 - [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -106,12 +106,14 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.Writer;
+import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -6120,7 +6122,7 @@ public abstract class AbstractJavadocMojo
 
             if ( location.exists() )
             {
-                String url = getJavadocLink( p );
+                String url = getApiDocLink( p );
 
                 OfflineLink ol = new OfflineLink();
                 ol.setUrl( url );
@@ -6173,13 +6175,14 @@ public abstract class AbstractJavadocMojo
 
                 if ( StringUtils.isNotEmpty( artifactProject.getUrl() ) )
                 {
-                    String url = getJavadocLink( artifactProject );
-
-                    if ( isValidJavadocLink( url, true ) )
+                    for ( String url: getJavadocLinks( artifactProject ) )
                     {
-                        getLog().debug( "Added Javadoc link: " + url + " for " + artifactProject.getId() );
+                        if ( isValidJavadocLink( url, true ) )
+                        {
+                            getLog().debug( "Added Javadoc link: " + url + " for " + artifactProject.getId() );
 
-                        dependenciesLinks.add( url );
+                            dependenciesLinks.add( url );
+                        }
                     }
                 }
             }
@@ -6518,12 +6521,23 @@ public abstract class AbstractJavadocMojo
     // ----------------------------------------------------------------------
 
     /**
+     * @return list of potential javadoc home urls to resolve <code>package-list</code> file
+     */
+    private static List<String> getJavadocLinks( MavenProject p )
+    {
+        return Arrays.asList(
+                getApiDocLink( p ),
+                getJavadocIOLink( p.getGroupId(), p.getArtifactId(), p.getVersion() )
+        );
+    }
+
+    /**
      * @param p not null
      * @return the javadoc link based on the project url i.e. <code>${project.url}/${destDir}</code> where
      *         <code>destDir</code> is configued in the Javadoc plugin configuration (<code>apidocs</code> by default).
      * @since 2.6
      */
-    private static String getJavadocLink( MavenProject p )
+    private static String getApiDocLink( MavenProject p )
     {
         if ( p.getUrl() == null )
         {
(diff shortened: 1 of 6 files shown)


maxcellent:
yeah, I've been thinking about the same when implementing it. But it feels
a bit overkill to add a generic mechanism if this is the only case.

How about merge this for now? and then we can further extends it if there
are more user requests to add more providers. Happy to help with the coding
as well.

eolivelli:
I am thinking about the mid term direction for the project.
Hardcoding javadoc site is okay, we are wrapping the javadoc tool...
Hardcoding a third party provider is something to think about a bit more.

Having a configurable list with a default list of known providers may be some kind of trade off.

I would like to know the option of other people in the community before moving forward with this work.
I think it is very useful and I would like to do it, but following the best way for the community

maxcellent:
sure. With regards to 

> I would like to know the option of other people

anything needs to be done from my side? or we will just wait until you get back to us?

eolivelli:
We are just waiting for other committers in the community to review this patch.
If noone objects I will move forward.
I would prefer to have configurable providers.
Cc @rfscholte

maxcellent:
hi @eolivelli , how are we going with the review?

rfscholte:
We're not going to include some third party URL's, so I agree with @eolivelli that this should be solved with providers.

maxcellent:
ok. I respect this.

By providers do you mean something like this?


   1. <configuration>
   2. <providers>
   3. <provider>http://javadoc.io/doc/[groupId]/[artifactId]/[versionId]
   </provider>
   4. <provider>...some other providers...</provider>
   5. </providers>
   6. ...
   7. </configuration>


Max

eolivelli:
Yes

elharo:
Can we close this PR then?

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -6517,13 +6520,24 @@ private boolean isJavadocVMInitError( String output )
     // Static methods
     // ----------------------------------------------------------------------
 
+    /**
+     * @return list of potential javadoc home urls to resolve <code>package-list</code> file
```
elharo, line 6524: urls --> URLs

should specify that this is returned in order of preference and indicate where URLs are looked up

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -6543,6 +6557,38 @@ private static String getJavadocLink( MavenProject p )
...
+     * @return the javadoc link based on the project javadoc.io url
+     *         i.e. <code>https://www.javadoc.io/doc/${group_id}/${artifact_id}/${version_id}/package-list</code>
+     */
+    static String getJavadocIOLink( String groupId, String artifactId, String version )
+    {
+        String baseUrl = "https://javadoc.io/doc/";
+        return baseUrl + urlEncode( groupId ) + "/" + urlEncode( artifactId ) + "/" + urlEncode( version );
+    }
+
+    static String urlEncode( String s )
```
elharo, line 6573: private

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -6543,6 +6557,38 @@ private static String getJavadocLink( MavenProject p )
...
     }
 
+    /**
+     * @param groupId not null
+     * @param artifactId not null
+     * @param version not null
+     * @return the javadoc link based on the project javadoc.io url
+     *         i.e. <code>https://www.javadoc.io/doc/${group_id}/${artifact_id}/${version_id}/package-list</code>
+     */
+    static String getJavadocIOLink( String groupId, String artifactId, String version )
```
elharo, line 6567: private

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -6543,6 +6557,38 @@ private static String getJavadocLink( MavenProject p )
...
+    {
+        String baseUrl = "https://javadoc.io/doc/";
+        return baseUrl + urlEncode( groupId ) + "/" + urlEncode( artifactId ) + "/" + urlEncode( version );
+    }
+
+    static String urlEncode( String s )
+    {
+        if ( s == null )
```
elharo, line 6575: don't need this check if your nonnull claims above are true. also this method should never return null

src/test/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojoTest.java
```diff
@@ -87,4 +88,20 @@ public void testMJAVADOC527_DetectLinksRecursion()
         assertFalse( mojo.isValidJavadocLink( "http://javamail.java.net/mailapi/apidocs", false ) );
         assertTrue( mojo.isValidJavadocLink( "http://commons.apache.org/proper/commons-lang/apidocs", false ) );
     }
+
+    public void test_urlEncode()
```
elharo, line 92: Let's try not to make a general URL encoding utility and test through the public API. 

There are better libraries to encode than URLEncoder though I'm not sure which are in our dependency tree already.

src/test/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojoTest.java
```diff
@@ -87,4 +88,20 @@ public void testMJAVADOC527_DetectLinksRecursion()
...
+    {
+        assertEquals("test+abc", urlEncode("test abc"));
+    }
+
+    public void test_urlEncode_null_input()
+    {
+        assertNull(urlEncode(null));
+    }
+
+    public void test_getJavadocIOLink()
```
elharo, line 102: testGetJavadocIOLink

rfscholte:
it shouldn't be too hard to solve this properly. I simply don't like the idea of overrides from the original javadoc locations to a hardcoded third party location.

rfscholte:
This looks like something we cannot agree on. So you're forcing me to pick this up myself. Adding such third party domains to our codebase is just not acceptable for me.

maxcellent:
I would love to see this perfect flexible solution. But it's a lot more effort than the current one so I haven't got a chance to work on that yet. Additionally there are few things need to be decided by the community first: 
- what would be the pattern? Do we need to define a DSL like this? `https://javadoc.io/doc/[group_id]/[artifact_id]/[version_id]`? 
- which 3rd party library is allowed?
- or @rfscholte when you say `it shouldn't be too hard` are you suggesting something simpler?

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
maxcellent: This has a default modifier so that unit test can test against it directly.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
maxcellent: This has a default modifier so that unit test can test against it directly.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
elharo: By testing like that you make it harder to change the implementation. Aside from the unrreachable catch block, you can test this code using only the public API.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
elharo: It can be tested through the public API.

rfscholte:
Closing this PR as this is based on the fixed third party URL. Will need to work on a dynamic solution.

jira-importer:
Resolve #920

PR #45, opened 2020-03-28
Uses try with resources

elharo:
@michael-o Starting work on clearing out some old code so we can update plexus and aether. Also fixes some API docs


elharo:
Build passing: https://builds.apache.org/job/maven-box/job/maven-javadoc-plugin/job/plexus/

PR #48, opened 2020-05-29
[MJAVADOC-653] fix javadoc; fix code smells

XenoAmess:


src/main/java/org/apache/maven/plugins/javadoc/AbstractFixJavadocMojo.java
```diff
@@ -808,20 +808,6 @@ private void parseClirrTextOutputFile( File clirrTextOutputFile )
                 switch ( code )
                 {
                     case 7011:
```
elharo, line 810: fallthrough is dangerous; perhpas extrat a helper method instead

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
```diff
@@ -629,35 +650,34 @@ protected static String parseJavadocMemory( String memory )
             throw new IllegalArgumentException( "The memory could not be null." );
         }
 
-        Pattern p = Pattern.compile( "^\\s*(\\d+)\\s*?\\s*$" );
+        Pattern p = PARSE_JAVADOC_MEMORY_PATTERN_0;
```
elharo, line 653: renaming this MEMORY_PATTERN_N does not help, arguably it makes this worse by making one look up the pattern

XenoAmess: @elharo basic idea is the original code compile the pattern every time it invoke, but actually we need compile it only one time.
after compile, it can be used thus do not need to compile it again when next try.

XenoAmess:
> The Travis CI should be pulled out to a separate issue and PR.

@elharo done. commits about travis-ci has been rebased out.

src/main/java/org/apache/maven/plugins/javadoc/AbstractFixJavadocMojo.java
XenoAmess: > fallthrough is dangerous; perhpas extrat a helper method instead

@elharo OK.

XenoAmess: @elharo done.

src/main/java/org/apache/maven/plugins/javadoc/AbstractFixJavadocMojo.java
```diff
@@ -856,6 +831,23 @@ private void parseClirrTextOutputFile( File clirrTextOutputFile )
...
 
+    private void methodAdded( String[] split )
+    {
+        String[] splits2;
+        List<String> list = clirrNewMethods.get( split[2].trim() );
+        if ( list == null )
+        {
+            list = new ArrayList<>();
+        }
+        splits2 = StringUtils.split( split[3].trim(), "'" );
```
elharo, line 842: initialize this here

src/test/java/org/apache/maven/plugins/javadoc/FixJavadocMojoTest.java
```diff
@@ -22,10 +22,7 @@
 import java.io.File;
 import java.io.IOException;
 import java.io.StringReader;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Properties;
+import java.util.*;
```
elharo, line 25: I'm not sure if apache maven has a style rule about this or not, but I know at Google we don't use wildcard imports.

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
```diff
@@ -629,35 +650,34 @@ protected static String parseJavadocMemory( String memory )
...
-        p = Pattern.compile( "^\\s*(\\d+)\\s*k(b)?\\s*$", Pattern.CASE_INSENSITIVE );
+        p = PARSE_JAVADOC_MEMORY_PATTERN_1;
         m = p.matcher( memory );
         if ( m.matches() )
         {
             return m.group( 1 ) + "k";
         }
 
-        p = Pattern.compile( "^\\s*(\\d+)\\s*m(b)?\\s*$", Pattern.CASE_INSENSITIVE );
+        p = PARSE_JAVADOC_MEMORY_PATTERN_2;
```
elharo, line 666: can you just use the constants here rather than assigning and reusing a local variable?

src/main/java/org/apache/maven/plugins/javadoc/AbstractFixJavadocMojo.java
XenoAmess: @elharo done.

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
XenoAmess: @elharo done.

src/test/java/org/apache/maven/plugins/javadoc/FixJavadocMojoTest.java
XenoAmess: @elharo done.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -6066,23 +6066,15 @@ private URL getResource( final List<String> classPath, final String resource )
                 getLog().error( "MalformedURLException: " + e.getMessage() );
             }
         }
-
-        URLClassLoader javadocClassLoader = new URLClassLoader( urls.toArray( new URL[urls.size()] ), null );
-        try
+        URL resourceURL = null;
```
elharo, line 6069: pull this and the return into the try block

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -6066,23 +6066,15 @@ private URL getResource( final List<String> classPath, final String resource )
...
-        URLClassLoader javadocClassLoader = new URLClassLoader( urls.toArray( new URL[urls.size()] ), null );
-        try
+        URL resourceURL = null;
+        try ( URLClassLoader javadocClassLoader = new URLClassLoader( urls.toArray( new URL[urls.size()] ), null ) )
         {
-            return javadocClassLoader.getResource( resource );
-        }
-        finally
+            resourceURL = javadocClassLoader.getResource( resource );
+        } catch ( IOException e )
```
elharo, line 6073: DANGER: this must be rethrown, not ignored

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
```diff
@@ -629,36 +650,31 @@ protected static String parseJavadocMemory( String memory )
...
-        Matcher m = p.matcher( memory );
+        Matcher m = PARSE_JAVADOC_MEMORY_PATTERN_0.matcher( memory );
         if ( m.matches() )
         {
             return m.group( 1 ) + "m";
         }
 
-        p = Pattern.compile( "^\\s*(\\d+)\\s*k(b)?\\s*$", Pattern.CASE_INSENSITIVE );
-        m = p.matcher( memory );
+        m = PARSE_JAVADOC_MEMORY_PATTERN_1.matcher( memory );
```
elharo, line 659: in general, we should avoid reusing local variables for different objects

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
```diff
@@ -1815,7 +1831,7 @@ private static CloseableHttpClient createHttpClient( Settings settings, URL url
         builder.setUserAgent( "Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.0)" );
 
         // Some server reject requests that do not have an Accept header
-        builder.setDefaultHeaders( Arrays.asList( new BasicHeader( HttpHeaders.ACCEPT, "*/*" ) ) );
+        builder.setDefaultHeaders( Collections.singletonList( new BasicHeader( HttpHeaders.ACCEPT, "*/*" ) ) );
```
elharo, line 1834: stet. Immutability makes Collections.singletonList not semantically equivalent to Arrays.asList

src/test/java/org/apache/maven/plugins/javadoc/AggregatorJavadocReportTest.java
```diff
@@ -173,20 +173,13 @@ private static String readFile( File file )
     {
         String strTmp;
```
elharo, line 174: could pull this into the try block or even the loop initializer

src/test/java/org/apache/maven/plugins/javadoc/JavadocJarTest.java
```diff
@@ -154,7 +154,7 @@ public void testInvalidDestdir()
         //check if the javadoc jar file was generated
         File generatedFile = new File( getBasedir(),
                                        "target/test/unit/javadocjar-invalid-destdir/target/javadocjar-invalid-destdir-javadoc.jar" );
-        assertTrue( !FileUtils.fileExists( generatedFile.getAbsolutePath() ) );
+        assertFalse(FileUtils.fileExists(generatedFile.getAbsolutePath()));
```
elharo, line 157: Maven style uses spaces before ) and after ( unless there are no arguments

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
XenoAmess: @elharo This IOException is ignored in the original codes, so I don't think we shall rethrow it...

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
XenoAmess: > Immutability makes Collections.singletonList not semantically equivalent to Arrays.asList
Yes but I don't think a immutable collection used here can be any problem.
I see no change to this collection after passed in as parameter here(in the codes).
Or please give me an example where can it break.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
XenoAmess: @elharo done.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
elharo: The ignored IOException was from closing in the finally block, not from other sources.

src/test/java/org/apache/maven/plugins/javadoc/AggregatorJavadocReportTest.java
XenoAmess: @elharo done.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
XenoAmess: @elharo no other resource will throw IOException here.
`public URL getResource(String name) {`
IOException is not a RuntimeException. If It can be thrown it must imply it.

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
XenoAmess: @elharo done.
btw, this reusing is by original codes, and I don't like the reuse either.
now fixed.

src/test/java/org/apache/maven/plugins/javadoc/JavadocJarTest.java
XenoAmess: @elharo done.
btw, is there a code formatter / style checker for this code for this format we used here?

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
elharo: There's no catch block in the old code. Maybe this isn't thrown at all?

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
elharo: The API accepts a list, not an immutable list, and does not guarantee that it will not try to mutate that list.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
XenoAmess: @elharo It seems only can throw when closing the `javadocClassLoader`. the original codes igonre it in `finally`. the new codes ignore it in `catch`. same thing IMO.

elharo: Then try with resources isn't working right here somehow. You shouldn't need to catch IOException that's only thrown on autoclose.

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
XenoAmess: @elharo fine, will revert this.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
XenoAmess: > Then try with resources isn't working right here somehow. You shouldn't need to catch IOException that's only thrown on autoclose.
@elharo 
Well, kind of agreed, but that is oracle guys' dicision for forcing us to must catch it...
Or should we revert it back to normal try instead of try-with-resource?

elharo: There's something I'm not seeing here. Have you tried simply removing the catch block completely? Does that compile?

XenoAmess: > There's something I'm not seeing here. Have you tried simply removing the catch block completely? Does that compile?

after removing that part:
```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  12.622 s
[INFO] Finished at: 2020-06-20T22:46:09+08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project maven-javadoc-plugin: Compilation failure
[ERROR] /D:/workspace/maven-javadoc-plugin/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java:[6068,30] unreported exception java.io.IOException; must be caught or declared to be thrown
[ERROR]   exception thrown from implicit call to close() on resource variable 'javadocClassLoader'
[ERROR] 
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException

```

elharo: Weird. I still don't see why. For now though, just leave the code as it was since autocloseable isn't helping

XenoAmess: @elharo done.

src/test/java/org/apache/maven/plugins/javadoc/JavadocJarTest.java
slachiewicz: please check here: https://maven.apache.org/developers/conventions/code.html

XenoAmess: > please check here: https://maven.apache.org/developers/conventions/code.html

@slachiewicz Thanks.
btw is there anybody for more reviewings to this pr?

XenoAmess:
> Can you put numbers on the performance improvement?

@elharo sorry but I didn't get what you mean by "put numbers on".
If you need performance test then I think I can try make some jmh tests.
But if you need real-time usecases performance test then I can't help with that.

elharo:
Absent measurement, it can't be claimed that a PR improves performance; and measuring these things is hard.

XenoAmess:
> Absent measurement, it can't be claimed that a PR improves performance; 

OK, if you see it this way I will not oppose.

> and measuring these things is hard.

Indeed. A good usecase for performance test can be very hard to built...

XenoAmess:
@elharo 
Hi.
What should we do next for this pr?
Should I sqruash all commits now?

XenoAmess:
> Yes, if you could rebase and squash your commits that would be wonderful. Thanks.

@elharo 
rebase done.

elharo:
Running it through Jenkins now: https://ci-builds.apache.org/job/Maven/job/maven-box/job/maven-javadoc-plugin/job/MJAVADOC-653/

XenoAmess:
@elharo 
Thanks.
As the codes are clean now, I can continue to next step.
The next pr will be some fix about some bugs in javadoc:fix .
I will try to do it as fast as possible.
Will create pr when I finish.

elharo:
Please file a JIRA issue for each actual bug.

XenoAmess:
> Please file a JIRA issue for each actual bug.

OK.

jira-importer:
Resolve #1015

PR #51, opened 2020-07-26
[MJAVADOC-623] - Add element-list for JDK 15

anthonyvdotbe:
This allows to build with a JDK for which the plugin does not contain a `java-api-element-list-XX` file yet.

## what
When building with a JDK version XX, for which the plugin does not contain a `java-api-element-list-XX` file yet, the plugin falls back to downloading the element-list from the actual `javaApiLink`.

## how
n/a

## why
Otherwise [this `if` statement](https://github.com/apache/maven-javadoc-plugin/blob/master/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java#L6410) is false, no `api-element-list` file is copied, the error from [MJAVADOC-623](https://issues.apache.org/jira/browse/MJAVADOC-623) occurs, and the Javadoc doesn't contain any links.

## checklist

       pulling in other changes.
       commit message.
 - [x] Run `mvn clean verify -Prun-its` to make sure basic checks pass. A more thorough check will 


src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -6407,20 +6410,41 @@ else if ( javaApiversion.asMajor().isAtLeast( "9" ) )
...
+            String elementListLink = javaApiLink + "element-list";
+            try
             {
-                // TODO only copy when changed
-                Files.copy( closableIS, javaApiListFile, StandardCopyOption.REPLACE_EXISTING );
+                URL elementListUrl = new URL( elementListLink );
+                String elementListContent = (String) elementListUrl.getContent( new Class<?>[] { String.class } );
+                if ( elementListContent != null )
+                {
+                    in = new ByteArrayInputStream( elementListContent.getBytes( UTF_8 ) );
```
eolivelli, line 6422: Why are you creating one String and one byte[] ?
What about simply assigning `is` to `elementListUrl.getInputStream` ?

anthonyvdotbe: Thanks for the review, that's much better indeed.

rfscholte:
Please revert all changes of `AbstractJavadocMojo.java`, better use `InputStream` ASAP, not via a URL.

anthonyvdotbe:
There are 2 changes in `AbstractJavadocMojo.java` to improve the case where the element-list is not embedded in the plugin:
1. fall back to the URL `javaApiLink + "element-list"`. This allows the plugin to work, regardless of which `element-list` files are embedded
2. log an error message and return `null`. This fixes the reported issue properly

W.r.t. `better use InputStream ASAP, not via a URL`: I believe the current patch is cleaner, since otherwise you'd need the same logic twice, once in case the embedded file is used, and once in case the fallback is used.

More generally, why are you against these changes? Once the plugin is updated to use the JDK mechanism instead, this whole method will be deleted anyway. And until then, it allows the plugin to work with EA builds of JDK 16.

rfscholte:
There's no need for any fallback. Let's ignore 16-ea, at a certain moment there will be a 16-ea that contains the automatic linking.

anthonyvdotbe:
Ok, I updated the PR.

rfscholte:
Merged with https://github.com/apache/maven-javadoc-plugin/commit/7a1b7fb5175c8ca39374a38fc91553ac4561877d

jira-importer:
Resolve #914

PR #52, opened 2020-08-20
[MJAVADOC-657] found a bug about wrongly handling annotations when using fix operation

XenoAmess:


XenoAmess:
BTW this bug was found when I build my lib https://github.com/cyanpotion/x8l
So yes, this bug does not only exist, but there really be users affected by it.

XenoAmess:
@elharo

XenoAmess:
@elharo Hi. I fixed the bug.
please review this pr.
thanks.

XenoAmess:
I really hope we can solve this fast.
Because I have another bug to fix for this repo. (that is also about "fix" operation)
But I don't  want to open two prs for this repo in same time (and on the same function)

src/test/resources/unit/fix-test/expected/src/main/java/fix/test/ClassWithJavadoc.java
```diff
@@ -44,6 +45,25 @@ public ClassWithJavadoc( String aString )
     {
     }
 
+    /**
+     * public constructor with annotation
+     *
+     * @param b a {@link java.lang.Boolean} object.
```
elharo, line 51: You don't need these tags in test resources, though the above comment is helpful

elharo:
Please file a JIRA issue for the second bug.

XenoAmess:
@elharo 
> Please file a JIRA issue for the second bug.

I fired that bug at MJAVADOC-658.

src/test/resources/unit/fix-test/expected/src/main/java/fix/test/ClassWithJavadoc.java
XenoAmess: @elharo
Yes, these functions are modified from my lib's functions where I meet this bug, I just forgot to delete the `@see` tags.
deleted now.

jira-importer:
Resolve #1034

PR #58, opened 2020-08-25
MJAVADOC-661: Fix for tagletArtifact fails to scan multi-release JARs (MRJARs) for Taglets

allen-ball:
JavadocUtil.getClassNamesFromJar(File) translates JarEntry names to Class names by first testing if the entry ends in ".class" and then translating the whole JarEntry name simply by replacing "/" with ".".  This fails for classes in multi-release JARs under the META-INF/versions/** hierarchy.

This change filters "*.class" files from the META-INF/versions/** hierarchy testing against JavaVersion.JAVA_VERSION.isAtLeast(String).

I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/src/it/projects/MJAVADOC-661_mrjar/invoker.properties
+++ b/src/it/projects/MJAVADOC-661_mrjar/invoker.properties
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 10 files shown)


elharo (changes requested):
This needs tests.

rfscholte:
Merged with https://github.com/apache/maven-javadoc-plugin/commit/981ac787ef2339dca42cbd2fe624a013bf5f810e

PR #60, opened 2020-09-10
[MJAVADOC-650] -  Fix Javadoc search path for Java 9+

wardev:
Fixes https://issues.apache.org/jira/browse/MJAVADOC-650
See discussion there.

Imports patch from Debian here: https://sources.debian.org/data/main/m/maven-javadoc-plugin/3.0.1-3/debian/patches/openjdk-9-javadoc-path.patch

Then fixes Javadoc.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -3783,8 +3783,35 @@ public abstract class AbstractJavadocMojo
 
         // ----------------------------------------------------------------------
         // Try to find javadocExe from System.getProperty( "java.home" )
-        // By default, System.getProperty( "java.home" ) = JRE_HOME and JRE_HOME
-        // should be in the JDK_HOME
+        // "java.home" is the "Installation directory for Java Runtime
+        // Environment (JRE)" used to run this code. I.e. it is the parent of
+        // the directory containing the `java` command used to start this
+        // application. It does not necessarily have any relation to the
+        // environment variable JAVA_HOME.
+        //
+        // In Java 8 and below the JRE is separate from the JDK. When
+        // installing the JDK to my-dir/ the javadoc command is installed in
+        // my-dir/bin/javadoc, the JRE is installed to my-dir/jre, and hence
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -3799,7 +3799,7 @@ else if ( SystemUtils.IS_OS_MAC_OSX && !JavaVersion.JAVA_SPECIFICATION_VERSION.i
         {
             javadocExe = new File( SystemUtils.getJavaHome() + File.separator + "bin", javadocCommand );
         }
-        // On Java 9, this has moved to: /usr/lib/jvm/java-9-openjdk-amd64/bin/javadoc
+        // On Java 9 the jre subdirectory was removed from the JDK
```
bmhm, line 3802: I am not a maintainer, but I think this comment could still go a bit more into details.

"On Java 9 the jre subdirectory was removed from the JDK" does not explain why the relative path from javaHome is different. The interesting part (as explained in the issue) is:
If the env_var `JAVA_HOME` was not set, and the java binary was called from a JDK installation from the jre subfolder, "java.home" was previously set to `JDK_HOME/jre`. Now as the jre directory disappeared, it is more obvious why this change is necessary.

Maybe a maintainer can tell what should go here.

wardev: It was intended to be an addendum to the comment about 15 lines above above, which I think includes the details your're looking for:

```
        // Try to find javadocExe from System.getProperty( "java.home" )
        // By default, System.getProperty( "java.home" ) = JRE_HOME and JRE_HOME
        // should be in the JDK_HOME
```

I can add a longer comment if you think it is worthwhile.
```

eolivelli: @wardev please add a longer comment as requested by @bmhm 
I feel it is clearer for future changes, when Java < 9 will be history

bmhm:
LGTM!

slachiewicz (approved):

PR #63, opened 2020-10-02
[MJAVADOC-664] - Fix stale javadoc data detection on Windows.

Syquel:
Replaces the usage of the full commandline by only its arguments to generate the stale javadoc detection key.
This approach allows to handle file-arguments appropriately in case of double-quoting by resolving them to their actual arguments.


michaelboyles:
Not a committer, but I think this unlikely to be accepted without a test

slachiewicz:
How I can test that functionality?

Syquel:
@slachiewicz I documented that mostly in [MJAVADOC-664](https://issues.apache.org/jira/browse/MJAVADOC-664).
Do you need anything else?

Writing a unit test for this fix would require some time, which I am currently not able to invest, because there is no unit test for the feature itself yet.

slachiewicz (approved):

jira-importer:
Resolve #1043

PR #77, opened 2021-05-06
[MJAVADOC-642] do not try to read from links in offline mode.

bmarwell:
- output error stack traces in debug mode.

Pulled out a trimmed-lowercase-link variable for readability and added a null check which oddly occured after creating this PR.
Please note that not even the non-integration-tests can be run behind a proxy.

       pulling in other changes.
       commit message.

 - [X] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -6686,14 +6686,19 @@ public abstract class AbstractJavadocMojo
      */
     protected boolean isValidJavadocLink( String link, boolean detecting )
     {
+        if ( link == null )
+        {
+          return false;
+        }
+
         try
         {
             final URI packageListUri;
             final URI elementListUri;
 
-            if ( link.trim().toLowerCase( Locale.ENGLISH ).startsWith( "http:" ) || link.trim().toLowerCase(
-                Locale.ENGLISH ).startsWith( "https:" ) || link.trim().toLowerCase( Locale.ENGLISH ).startsWith(
-                "ftp:" ) || link.trim().toLowerCase( Locale.ENGLISH ).startsWith( "file:" ) )
+          final String trimmedLowercaseLink = link.trim().toLowerCase( Locale.ENGLISH );
+          if ( trimmedLowercaseLink.startsWith( "http:" ) || trimmedLowercaseLink.startsWith( "https:" )
+              || trimmedLowercaseLink.startsWith( "ftp:" ) || trimmedLowercaseLink.startsWith( "file:" ) )
             {
                 packageListUri = new URI( link + '/' + PACKAGE_LIST );
                 elementListUri = new URI( link + '/' + ELEMENT_LIST );
@@ -6721,6 +6726,11 @@ public abstract class AbstractJavadocMojo
                 elementListUri = new File( dir, ELEMENT_LIST ).toURI();
             }
 
+            if ( isOffline && !trimmedLowercaseLink.startsWith( "file:" ) )
+            {
+              // checking of both elementlists and package lists skipped.
+              return false;
+            }
 
             try
             {


rfscholte:
Please don't split the check for package-list and element-list. If there's no element-list you shouldn't log, as the package-list could exist. Only if oth are missing, there's a reason to log.

rfscholte (changes requested):
combine package-list/element-list message to prevent unnecesary noise in logging

bmarwell:
> combine package-list/element-list message to prevent unnecesary noise in logging

Will do (revert). I thought it might have been useful for debugging (-X), but this is probably out of scope for this PR.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -6677,6 +6682,11 @@ protected boolean isValidJavadocLink( String link, boolean detecting )
                 elementListUri = new File( dir, ELEMENT_LIST ).toURI();
             }
 
+            if ( isOffline && !trimmedLowercaseLink.startsWith( "file:" ) )
+            {
+              // checking of both elementlists and package lists skipped.
+              return false;
```
rfscholte, line 6688: `localhost` and `127.0.0.1` are considered safe too. Makes we wonder if there's a way to use Mavens own Wagon instance that already respects the configuration of settings.xml

bmarwell: @rfscholte do you have any explicit suggestion how to resolve this? Technically "offline" could also mean "no network", where 127.0.0.1 is a local network interface, too. But if there is a better solution, I would be willing to implement it.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -1457,21 +1457,21 @@
...
+     * </pre>
      * @since 3.3.0
-     */
-    @Parameter
-    private String[] addStylesheets;
-    
+     */
+    @Parameter
+    private String[] addStylesheets;
+
```
bmarwell, line 1474: Mixed windows linebreaks converted to unix by IDE.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -3119,34 +3119,34 @@ private void addAddStyleSheets( List<String> arguments ) throws MavenReportExcep
...
+                                                        + "than stylesheetfile: " + stylesheetfile.get().getName() );
+                }
+            }
+            
+            return Optional.of( addstylesheetfile );
+        }
+
+        throw new MavenReportException( "additional stylesheet file does not exist: " 
+                                            + addstylesheetfile.getAbsolutePath() );
+    }
```
bmarwell, line 3149: Mixed windows linebreaks converted to unix by IDE.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -5643,7 +5643,7 @@ private void addStandardDocletOptions( File javadocOutputDirectory,
                               JavadocUtil.quotedPathArgument( stylesheetfile.get().getAbsolutePath() ) );
         }
         
-        addAddStyleSheets( arguments ); 
+        addAddStyleSheets( arguments );
```
bmarwell, line 5646: Mixed windows linebreaks converted to unix by IDE.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
bmarwell: Well, a quick interim solution could be sth like this:

```diff
Index: src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java	(revision 92957646c0be81474fcbf1e84032880f2e2b350e)
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java	(date 1621452882333)
@@ -4368,7 +4368,7 @@
                 continue;
             }
 
-            if ( isOffline && !link.startsWith( "file:" ) )
+            if ( isOffline && isOnlineLink( link ))
             {
                 continue;
             }
@@ -4382,6 +4382,19 @@
         }
     }
 
+    private boolean isOnlineLink( String link ) {
+        if  ( link.startsWith( "file:" ) ) {
+            return true;
+        }
+
+        if ( link.startsWith( "http" ) &&
+            (link.contains( "//localhost" ) || link.contains( "//127.0.0." ) ) ) {
+            return false;
+        }
+
+        return true;
+    }
+
     /**
      * Coppy all resources to the output directory
      *
@@ -6726,7 +6739,7 @@
                 elementListUri = new File( dir, ELEMENT_LIST ).toURI();
             }
 
-            if ( isOffline && !trimmedLowercaseLink.startsWith( "file:" ) )
+            if ( isOffline && isOnlineLink( trimmedLowercaseLink ))
             {
               // checking of both elementlists and package lists skipped.
               return false;
```

But I fear such provisional changes, as nothings holds as forever as provisionally solutions.

rfscholte:
When this plugin wants to use the settings' `offline`, it should behave exactly the same as Maven uses this property, which means Wagon should be used. This will completely replace all httpclient code. I can make a new ticket which will superseed this one if you want.

bmarwell:
Sure, go ahead. There's more to change my introducing wagon which would be out of scope in this issue.

cstamas:
my 5 cents: a big NO to use Wagon (maven2) stuff in any maven plugin, not is just an overkill, but wrong IMO as well...
Btw "it should behave exactly the same as Maven uses this property, which means Wagon should be used", does it means that the logic defining Maven behavior re offline is not in Maven but in Wagon?

rfscholte:
I should have said: the same implementation handling URL requests. Based of https://maven.apache.org/resolver/index.html that could still be Wagon. Based on the image we should be using the same "transport", right?

bmarwell:
> I should have said: the same implementation handling URL requests. Based of https://maven.apache.org/resolver/index.html that could still be Wagon. Based on the image we should be using the same "transport", right?

This PR is about LESS artefacts to transport. Why change the transport in such a PR?

rfscholte:
The transport layer is the central location to understand offline.

jira-importer:
Resolve #970

PR #106, opened 2021-10-28
[MJAVADOC-697] Allowing to register alternative Javadoc implementation

dukescript:
The Maven compiler plugin allows different implementations of compilers to plug in. One of the very useful ones is the [retrofit compiler for Java](http://frgaal.org) that allows one to use syntax of JDK-16 while running on old JDKs.

The compiler works fine, however the problem comes when one wants to generate Javadoc. Because the source code is using newer language constructs, classical javadoc from old JDKs fails. One either has to generate the Javadoc on a newest JDK (which beats the purpose of retrofit compiler) or give up on Javadoc. Or...

Let's enhance the Maven Javadoc Plugin to support alternative implementations of Javadoc, just like the Maven Compiler Plugin does with compilers!

 - [ x ] Addressing [MJAVADOC-697](https://issues.apache.org/jira/browse/MJAVADOC-697)
 - [ x ] Each commit in the pull request should have a meaningful subject line and body.
 - [ x ] Format the pull request title like `[MJAVADOC-XXX]` 
 - [ x ] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [ x ] Run `mvn clean verify -Prun-its` to make sure basic checks pass. A more thorough check will 

Pull request is about ~20 lines of code - no need to sign an [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf). To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0) you have to acknowledge this by using the following check-box.

 - [ x ] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -293,6 +293,9 @@ public abstract class AbstractJavadocMojo
     @Component
     private DependencyResolver dependencyResolver;
 
+    @Component
+    private List<JavadocImplementation> javadocInvocators;
+
     /**
      * Project builder
      *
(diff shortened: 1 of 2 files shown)


jira-importer:
Resolve #1072

PR #183, opened 2023-02-14
[MNG-6825] Replace org.apache.maven.shared.utils.StringUtils

timtebeek:
So from https://issues.apache.org/jira/browse/MNG-6825 it seems the goal is to reduce the dependency on `org.apache.maven.shared.utils.StringUtils`, and replace that with `org.apache.commons.lang3.StringUtils` when also a dependency. (Unless I misunderstood). This is a first step towards that, after which only `io.FileUtils` is holding up dropping `maven-shared-utils`.


timtebeek:
Closed while waiting on feedback in https://issues.apache.org/jira/browse/MNG-6825

PR #337, opened 2024-11-16
[MJAVADOC-814] handle parameters such packages with multi lines

olamy:
Signed-off-by: Olivier Lamy <olamy@apache.org>

 - [ ] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MJAVADOC) filed
       pulling in other changes.
 - [ ] Each commit in the pull request should have a meaningful subject line and body.
 - [ ] Format the pull request title like `[MJAVADOC-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.

 - [ ] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -5955,7 +5955,7 @@ public abstract class AbstractJavadocMojo extends AbstractMojo {
         options.setBootclasspathArtifacts(toList(bootclasspathArtifacts));
         options.setDocfilesSubdirsUsed(docfilessubdirs);
         options.setDocletArtifacts(toList(docletArtifact, docletArtifacts));
-        options.setExcludedDocfilesSubdirs(excludedocfilessubdir);
+        options.setExcludedDocfilesSubdirs(excludedocfilessubdir == null ? null : excludedocfilessubdir.trim());
         options.setExcludePackageNames(toList(excludePackageNames));
         options.setGroups(toList(groups));
         options.setLinks(links);
--- a/src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
@@ -57,6 +57,7 @@ import java.util.jar.JarInputStream;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
 
 import org.apache.http.HttpHeaders;
 import org.apache.http.HttpHost;
@@ -223,12 +224,12 @@ public class JavadocUtil {
     protected static String quotedArgument(String value) {
         String arg = value;
 
+        List<String> list =  Arrays.stream(arg.split("\n")).map(String::trim).collect(Collectors.toList());
+        arg = String.join("", list);
+
         if (arg != null && !arg.isEmpty()) {
             arg = arg.replace("'", "\\'");
             arg = "'" + arg + "'";
-
-            // To prevent javadoc error
-            arg = arg.replace("\n", " ");
         }
 
         return arg;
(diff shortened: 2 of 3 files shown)


src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -5955,7 +5955,7 @@ protected final JavadocOptions buildJavadocOptions() throws IOException {
         options.setBootclasspathArtifacts(toList(bootclasspathArtifacts));
         options.setDocfilesSubdirsUsed(docfilessubdirs);
         options.setDocletArtifacts(toList(docletArtifact, docletArtifacts));
-        options.setExcludedDocfilesSubdirs(excludedocfilessubdir);
+        options.setExcludedDocfilesSubdirs(excludedocfilessubdir == null ? null : excludedocfilessubdir.trim());
```
elharo, line 5958: This was hard to read. Consider renaming excludedDocfilesSubdirs

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
```diff
@@ -221,14 +221,17 @@ protected static List<String> getExcludedPackages(
      * @return quoted option-argument
      */
     protected static String quotedArgument(String value) {
+        if (value == null) {
+            return null;
+        }
         String arg = value;
 
+        List<String> list = Arrays.stream(arg.split("\n")).map(String::trim).collect(Collectors.toList());
+        arg = String.join("", list);
```
elharo, line 230: try to avoid reusing variables as it's bug prone

Also, this is combining lines with no space in between one line and the next. Is that what you want?

src/test/java/org/apache/maven/plugins/javadoc/JavadocUtilTest.java
```diff
@@ -675,4 +680,23 @@ private void stopSilently(Server server) {
             // ignored
         }
     }
+
+    public void testQuotedArgument() throws Exception {
+
+        String value = "      org.apache.uima.analysis_component:\n      org.apache.uima.analysis_engine\n";
+
+        String arg = JavadocUtil.quotedArgument(value);
+        assertThat(arg).isEqualTo("'org.apache.uima.analysis_component:org.apache.uima.analysis_engine'");
```
elharo, line 689: assertEquals is simpler and more direct

src/test/java/org/apache/maven/plugins/javadoc/JavadocUtilTest.java
```diff
@@ -675,4 +680,23 @@ private void stopSilently(Server server) {
...
+
+        value = "org.apache.uima.analysis_component:org.apache.uima.analysis_engine";
```
elharo, line 691: This should be two test methods so they don't interfere with eech other

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
olamy: I don't want to have to rename a Mojo parameter this will be non backward compact. https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#excludedocfilessubdir

src/test/java/org/apache/maven/plugins/javadoc/JavadocUtilTest.java
olamy: done via 74b03e61bbe839f1ab08f8ceb8594e1a6052a37f

src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
olamy: > try to avoid reusing variables as it's bug prone
> 
yup please look at the context of the code (the full code) yes it's not idea but it;s how the rest of the code is and I don't want to change everything this could be part of another PR if you like.

> Also, this is combining lines with no space in between one line and the next. Is that what you want?

jira-importer:
Resolve #1192

PR #350, opened 2024-12-05
[MJAVADOC-825] Prefer NullPointerExceptions for null arguments

elharo:


src/main/java/org/apache/maven/plugins/javadoc/JavadocUtil.java
```diff
@@ -534,11 +534,16 @@ protected static JavaVersion getJavadocVersion(File javadocExe)
      * @return the version of the javadoc for the output, only digits and dots
      * @throws PatternSyntaxException if the output doesn't match with the output pattern
      *             {@code (?s).*?[^a-zA-Z]([0-9]+\\.?[0-9]*)(\\.([0-9]+))?.*}.
-     * @throws IllegalArgumentException if the output is null
+     * @throws NullPointerException if the output is null
+     * @throws IllegalArgumentException if the output is empty
      */
     protected static String extractJavadocVersion(String output) throws IllegalArgumentException {
```
pzygielo, line 540: Given `IAEx` being declared as thrown by, perhaps `NPEx` could also be added for style consistency?
Or `IAEx` declaration removed?

elharo: Both have @throws clauses. Or did you mean something else?

pzygielo: `IllegalArgumentException` is `RTEx` (as `NPEx` is). And is declared as being thrown by this method. Thus - let's add `NullPointerException` to the `throws` list.
Or remove `throws`.
Here, and in the second modified method.

pzygielo: I see this:
```java
protected static String extractJavadocVersion(String output) throws IllegalArgumentException {
```
I'd like to see this instead:
```java
protected static String extractJavadocVersion(String output) throws IllegalArgumentException, NullPointerException {
```
or this:
```java
protected static String extractJavadocVersion(String output) {
```

elharo: Good spot. That should go the other way per guidelines. Removed both.

michael-o (approved):

PR #365, opened 2025-04-19
Enable GH issues

Bukama:


.github/workflows/release-drafter.yml
```diff
@@ -20,6 +20,8 @@ on:
   push:
     branches:
       - master
+workflow_dispatch:
```
pzygielo, line 23: I'd think this needs to be on the same level as `push`.

Bukama: Good spot, thank you.

Bukama:
PR template updated (as requested in dependency-plugin PR)

slawekjaranowski (approved):

slawekjaranowski (approved):

PR #380, opened 2025-05-30
Remove some log junk

elharo:
Reduce verbosity of logs


michael-o (changes requested):
I think that we can deprecate this parameter now. If someone really uses such an old version then it is their problem.

PR #385, opened 2025-05-30
Use Java 7 relativization instead of hand-rolled code

elharo:
There are actually two different toRelative methods used in this class, probably added at different times by different developers. This PR removes one of them.


elharo:
CI is flaky and broken

PR #393, opened 2025-06-05
Remove deprecated DefaultArtifactRepository dependency we don't need

elharo:


michael-o (approved):
Is there is no use, remove it!

PR #1215, opened 2025-07-12
Be consistent about data encoding when copying files

fridrich:
The file-name encoding should be consistent through the StaleHelper. If one runs the build in a non-english environment, where UTF-8 is not a default encoding, this can result in error while writing the javadoc. This commit fixes it.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/StaleHelper.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/StaleHelper.java
@@ -41,6 +41,20 @@ import org.codehaus.plexus.util.cli.Commandline;
 public class StaleHelper {
 
     /**
+     * Compute the encoding of the stale javadoc
+     *
+     * @return the the encoding of the stale data
+     */
+    private static Charset getDataCharset() {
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/plugins/javadoc/StaleHelper.java
```diff
@@ -55,13 +69,7 @@ public static List<String> getStaleData(Commandline cmd) throws MavenReportExcep
...
             Collections.addAll(options, args);
 
-            final Charset cs;
-            if (JavaVersion.JAVA_SPECIFICATION_VERSION.isAtLeast("9")
-                    && JavaVersion.JAVA_SPECIFICATION_VERSION.isBefore("12")) {
-                cs = StandardCharsets.UTF_8;
-            } else {
-                cs = Charset.defaultCharset();
-            }
+            final Charset cs = getDataCharset();
```
olamy, line 72: Maybe the local variable is not even needed?

olamy:
Thanks for your contribution

PR #1217, opened 2025-07-12
Make the legacyMode consistent and actually useful

fridrich:
* Filter out all of the module-info.java files in legacy mode
  * Do not use --source-path in legacy mode as not to suck any of those module-info.java files back
  * Generate the javadoc from list of files and not list of packages, which is not working if --source-path is not specified


fridrich:
Any comments on this one?

PR #1219, opened 2025-07-14
Respect the SOURCE_DATE_EPOCH variable from environment for reproducible builds

fridrich:
If the SOURCE_DATE_EPOCH variable is not set in the environment, nothing should change. If it is set, the outputTimestemp is the first to consider, but if it is not set, or it is cancelled, in the case where the outputTimeStamp is set to SOURCE_DATE_EPOCH.
This environmental variable is basically the standard across linux distributors for reproducible builds, and it is unlikely to be set otherwise. But this is basically for me a weather balloon to check whether something like this could be accepted upstream. If not, I will continue to carry it as a patch in our builds.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -2671,6 +2671,15 @@ public abstract class AbstractJavadocMojo extends AbstractMojo {
     private String getBottomText() {
         final String inceptionYear = project.getInceptionYear();
 
+        final String sourceDateEpoch = System.getenv("SOURCE_DATE_EPOCH");
+        if (sourceDateEpoch != null) {
+            if (outputTimestamp == null
+                    || outputTimestamp.length() < 1
+                    || ((outputTimestamp.length() == 1) && !Character.isDigit(outputTimestamp.charAt(0)))) {
+                outputTimestamp = sourceDateEpoch;
+            }
+        }
+
         // get Reproducible Builds outputTimestamp date value or the current local date.
         final LocalDate localDate = MavenArchiver.parseBuildOutputTimestamp(outputTimestamp)
                 .map(instant -> instant.atZone(ZoneOffset.UTC).toLocalDate())


olamy:
> How well would something like this be received?

why not this sounds some documented env var. but the title of your PR might need some improvement :)

olamy:
Reading this documentation 
https://github.com/apache/maven-javadoc-plugin/blob/33c9f01af9a3d4d28decbabd0bc02c4b3a875c2d/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java#L1641

I wonder why you cannot simply use this in your pom:
```
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-javadoc-plugin</artifactId>
        <version>@project.version@</version>
        <configuration>
          <outputTimestamp>${env.SOURCE_DATE_EPOCH}</outputTimestamp>
        </configuration>
      </plugin>
```

my other concern here is to have nothing documented about this anywhere except in the code. 
what about a new mojo parameter, such:
```
    /**
     * see <a href="https://reproducible-builds.org/docs/source-date-epoch/">SOURCE_DATE_EPOCH</a>).
     *
     * @since xxx
     */
    @Parameter
    protected String sourceDateEpoch;
```
and usage could be:
```
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-javadoc-plugin</artifactId>
        <version>@project.version@</version>
        <configuration>
          <sourceDateEpoch>${env.SOURCE_DATE_EPOCH}</sourceDateEpoch>
        </configuration>
      </plugin>
```
or document it as if empty value from env var `SOURCE_DATE_EPOCH` will be used if available.
At least this feature will be documented

olamy:
@hboutemy this one might interest you.

src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
```diff
@@ -2668,6 +2668,15 @@ private void populateCompileArtifactMap(Map<String, Artifact> compileArtifactMap
     private String getBottomText() {
         final String inceptionYear = project.getInceptionYear();
 
+        final String sde = System.getenv("SOURCE_DATE_EPOCH");
```
Bukama, line 2671: Please use proper variable names

Bukama (changes requested):
Aside my comment above and the discussion about how to set the thing up I miss updates to the documentation.

fridrich:
> SOURCE_DATE_EPOCH is when a project is not reproducible by default (ie. only the distro packager cares, but not the upstream project)
> 
> when a project is reproducible by default (= upstream), it is done in `pom.xml` and follows https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=74682318
> 
> please don't regress: if you really want to not embedd RB in pom.xml (I'm interested to hear why...), just run inject the value on the command line if you really want: `mvn -Dproject.build.outputTimestamp=$SOURCE_DATE_EPOCH package`

As a linux distribution, we package ~900 Java packages. Yes, I know that I can specify the project.build.outputTimestamp on command-line and that is what we were doing. Just that it is extremely error-prone. That is why I came with this change that will not affect at all any of the projects that are already reproducible and specify the timestamp themselves. It is only affecting projects that are not reproducible by default. As you can see from the code. Moreover, if the project specifies the timestamp, it is preferred to the SOURCE_DATE_EPOCH. So, we are not imposing that. The only case when the fallback is used is when the project does not do anything around reproducibility themselves.

Now, I am carrying this patch in our repositories for more then a year and was reluctant to try to upstream it. Nevertheless, I was still thinking that it could be useful for the maven eco-system. For instance, other build systems like ant respect the SOURCE_DATE_EPOCH.

I was also thinking about the security implications of acting upon an environment variable. But reading the parsers of the outputTimestamp, I don't see anything that could go wrong with that since if the parsing is unsuccessful, it will just not be set.

PR #1233, opened 2025-08-13
Remove unneeded checks

elharo:
Mostly cases where a variable is guaranteed non-null due to prior reference or method annotated NotNull, and some reworks of logic so continue isn't needed


elharo:
Error:    FixJavadocMojoTest.testFix:109->executeMojoAndTest:509->invokeCompileGoal:542 » NullPointer Cannot invoke "java.io.File.exists()" because "javaHome" is null

PR #1238, opened 2025-08-15
protect 3.8.x branch

hboutemy:
typo in #1228

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -30,7 +30,7 @@ github:
     rebase: true
   protected_branches:
     master: { }
-    maven-javadoc-plugin-3.8.3: { }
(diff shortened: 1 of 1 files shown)


sparsick (approved):

.asf.yaml
```diff
@@ -30,7 +30,7 @@ github:
     rebase: true
   protected_branches:
     master: { }
-    maven-javadoc-plugin-3.8.3: { }
+    maven-javadoc-plugin-3.8.x: { }
```
olamy, line 33: Not sure we really plan to make new releases from this branch 
Protecting this is not really needed. Branch could be even deleted

Issue #1242, opened 2025-08-18
`package {...} does not exist` when upgrading `3.11.2` -> `3.11.3` + `legacyMode`


JackPGreen:
### Affected version

3.11.3

### Bug description

When upgrading from [`3.11.2` -> `3.11.3`](https://github.com/hazelcast/hazelcast/commit/8f8c1b32bf61ca2e32d8125256e6b92c8ee627e1) (with [`legacyMode` mode](https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html#legacyMode)), `javadoc:aggregate` fails with:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.11.3:jar (attach-javadocs) on project {project}: MavenReportException: Error while generating Javadoc: 
[ERROR] Exit code: 1
[ERROR] /path/to/Class.java:19: error: package com.my.package does not exist
[ERROR] import com.my.package.OtherClass;
```

https://github.com/apache/maven-javadoc-plugin/pull/1217 introduced this issue. Specifically I note:
> looks good but I need to test on large projects such Jetty mixing jpms and not jpms for aggregated javadoc.

Which the project falls into.

Reproducer:
- Clone [this commit](https://github.com/hazelcast/hazelcast/commit/8f8c1b32bf61ca2e32d8125256e6b92c8ee627e1)
- `mvn clean install javadoc:aggregate -Prelease-snapshot -DskipTests`

PR #1243, opened 2025-08-18
Fix `package {...} does not exist` in `legacyMode`

JackPGreen:
_Partially_ reverts change in https://github.com/apache/maven-javadoc-plugin/pull/1217, specifically:
> Do not use --source-path in legacy mode as not to suck any of those module-info.java files back

This is required to ensure correct package resolution on projects that mix modular & non-module Java code.

Fixes: https://github.com/apache/maven-javadoc-plugin/issues/1242

### DIFF
--- a/src/it/projects/GITHUB-1242/invoker.properties
+++ b/src/it/projects/GITHUB-1242/invoker.properties
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#   http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+invoker.goals=package javadoc:aggregate
--- a/src/it/projects/GITHUB-1242/pom.xml
+++ b/src/it/projects/GITHUB-1242/pom.xml
@@ -1,0 +1,37 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+~   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


JackPGreen:
> test really is mandatory for a bug like this

I understand the value but adding one is non-trivial - would need to slim down the reproducer to suit, I don't think any existing test covers this scenario (specifically - the `legacy` + multi-module case I saw has no `import`s). I'll try to revisit.

Also - the fact that this fix doesn't affect any existing tests but _does_ fix my issues suggests that (in hindsight) there was insufficient test coverage for the previous PR.

JackPGreen:
> test really is mandatory for a bug like this

@elharo I've added a test in e1a74738a2fbef8e191e31e580452deb51e8c918, based on a slimmed-down version of the reproducer.

This test fails on `master`:
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.11.4-SNAPSHOT:aggregate (default-cli) on project GITHUB-1242: An error has occurred in Javadoc report generation: 

But passes on my branch:
 > [INFO] Building: GITHUB-1242/pom.xml
[INFO]           GITHUB-1242/pom.xml .............................. SUCCESS (1.861 s)

olamy (approved):
LGTM. I have been tested this on a large project such https://github.com/jetty/jetty.project with `mvn clean install -Pjavadoc-aggregate javadoc:aggregate -DskipTests`
and all good

fridrich:
The problem of the sourcepath is that it sucks in the module-info.java files in aggregate goal. I am getting this for instance:
`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.11.3:aggregate (default-cli) on project angus-activation-project: An error has occurred in Javadoc report generation: 
[ERROR] Exit code: 1
[ERROR] /home/abuild/rpmbuild/BUILD/angus-activation-2.0.2-build/angus-activation-2.0.2/activation-registry/src/main/java/module-info.java:17: error: module not found: jakarta.activation
[ERROR]     requires transitive jakarta.activation;
[ERROR]                                ^
[ERROR] error: cannot access module-info
[ERROR]   cannot resolve modules
[ERROR] 2 errors
[ERROR] Command line was: /usr/lib64/jvm/java-21-openjdk-21/bin/javadoc -J-Duser.language= -J-Duser.country= @options @argfile
`

fridrich:
But what I tried also to avoid in the original commit was not to refer to the sources by package. Maybe I missed a place. If one does not use the package names in legacy mode, one does not need the sourcepath

fridrich:
What I am also puzzled about is that I revert your AbstractJavadocMojo.java change and run the `mvn verify` the IT tests pass without a glitch.

Issue #1273, opened 2025-10-10
Regression starting in 3.11.3: `java.nio.charset.MalformedInputException: Input length = 1`


garydgregory:
### Affected version

3.11.3, 3.12.0

### Bug description

The discussion thread on the user list is at https://lists.apache.org/thread/o0x077z13mpv4d81287d9r3oltp3sf50

Reproduce on Windows and Java 8 like https://github.com/apache/commons-lang/actions/runs/18404870422/job/52453057388:

```
git clone https://gitbox.apache.org/repos/asf/commons-lang.git
cd commons-lang
git checkout b70093574020cfe8af035e562e184d77e278fafe # this is where we are now
mvn clean javadoc:javadoc && mvn javadoc:javadoc
```
Start of CI build:
```
Run mvn --errors --show-version --batch-mode --no-transfer-progress -Ddoclint=all
Apache Maven 3.9.11 (3e54c93a704957b63ee3494413a2b544fd3d825b)
Maven home: C:\ProgramData\chocolatey\lib\maven\apache-maven-3.9.11
Java version: 1.8.0_462, vendor: Temurin, runtime: C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\8.0.462-8\x64\jre
Default locale: en_US, platform encoding: Cp1252
OS name: "windows server 2022", version: "10.0", arch: "amd64", family: "windows"
```
Full stack trace:
```
Error:  Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.12.0:jar (create-javadoc-jar) on project commons-lang3: Execution create-javadoc-jar of goal org.apache.maven.plugins:maven-javadoc-plugin:3.12.0:jar failed: java.nio.charset.MalformedInputException: Input length = 1 -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.12.0:jar (create-javadoc-jar) on project commons-lang3: Execution create-javadoc-jar of goal org.apache.maven.plugins:maven-javadoc-plugin:3.12.0:jar failed: java.nio.charset.MalformedInputException: Input length = 1
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:333)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
[...]

Issue #1264, opened 2025-09-04
Build failures after update to v3.11.3 - java.nio.charset.MalformedInputException: Input length = 1


ottlinger:
### Affected version

3.11.3

### Bug description

After updating the maven-javadoc-plugin there are build failures on the dependabot branches in GitHubActions - e.g. https://github.com/ottlinger/spamschutz/actions/runs/17065661919/job/49379112974?pr=496

As the error message is far too generic it's unclear which/if any file triggers the problem. Not upgrading the javadoc plugin keeps the builds green (v3.11.2).

`
Caused by: java.io.UncheckedIOException: java.nio.charset.MalformedInputException: Input length = 1
    at java.nio.file.FileChannelLinesSpliterator.readLine (FileChannelLinesSpliterator.java:192)
    at java.nio.file.FileChannelLinesSpliterator.forEachRemaining (FileChannelLinesSpliterator.java:132)
    at java.util.stream.AbstractPipeline.copyInto (AbstractPipeline.java:509)
    at java.util.stream.AbstractPipeline.wrapAndCopyInto (AbstractPipeline.java:499)
    at java.util.stream.ReduceOps$ReduceOp.evaluateSequential (ReduceOps.java:921)
    at java.util.stream.AbstractPipeline.evaluate (AbstractPipeline.java:234)
    at java.util.stream.ReferencePipeline.collect (ReferencePipeline.java:682)
    at org.apache.maven.plugins.javadoc.AbstractJavadocMojo.isUpToDate (AbstractJavadocMojo.java:5010)
    at org.apache.maven.plugins.javadoc.AbstractJavadocMojo.executeJavadocCommandLine (AbstractJavadocMojo.java:4988)
    at org.apache.maven.plugins.javadoc.AbstractJavadocMojo.executeReport (AbstractJavadocMojo.java:2058)
    at org.apache.maven.plugins.javadoc.JavadocJarMojo.doExecute (JavadocJarMojo.java:196)
    at org.apache.maven.plugins.javadoc.AbstractJavadocMojo.execute (AbstractJavadocMojo.java:1850)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:328)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
[...]

PR #1274, opened 2025-10-10
Fix MalformedInputException on Windows with Java 8 due to charset mis…

slachiewicz:
…match in stale data cache

## Problem

Starting in version 3.11.3, running `mvn javadoc:javadoc` twice in succession on Windows with Java 8 fails on the second run with:

```
java.nio.charset.MalformedInputException: Input length = 1
    at org.apache.maven.plugins.javadoc.AbstractJavadocMojo.isUpToDate(AbstractJavadocMojo.java:5008)
```

This regression affects Windows users running Java 8, where the default platform encoding is `Cp1252`.

## Root Cause

The issue stems from a charset mismatch in how the stale data cache file is written versus how it's read:

1. **Write operation** (`StaleHelper.writeStaleData()`, line 126):
   - Uses `getDataCharset()` which returns `Charset.defaultCharset()` for Java 8
   - On Windows, this is `Cp1252`

2. **Read operation** (`AbstractJavadocMojo.isUpToDate()`, line 5008):
   - Always uses hardcoded `StandardCharsets.UTF_8`

When the second run attempts to read a file written with `Cp1252` encoding using `UTF-8`, any non-ASCII bytes cause a `MalformedInputException`.

## Solution

Changed `StaleHelper` to always save in `StandardCharsets.UTF_8` instead of platform-dependent charset. This ensures:
- Consistent encoding across all platforms (Windows, macOS, Linux)
- Consistent encoding across all Java versions (8, 11, 17, 21, etc.)
- Write and read operations use the same charset

Fixes: #1273 #1264

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/StaleHelper.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/StaleHelper.java
@@ -20,6 +20,8 @@ package org.apache.maven.plugins.javadoc;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.charset.CharacterCodingException;
+import java.nio.charset.Charset;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -56,8 +58,12 @@ public class StaleHelper {
             for (String arg : args) {
                 if (arg.startsWith("@")) {
                     String name = arg.substring(1);
-                    options.addAll(Files.readAllLines(dir.resolve(name), UTF_8));
                     ignored.add(name);
+                    try {
+                        options.addAll(Files.readAllLines(dir.resolve(name), UTF_8));
+                    } catch (CharacterCodingException e) {
+                        options.addAll(Files.readAllLines(dir.resolve(name), Charset.defaultCharset()));
+                    }
                 }
             }
             List<String> state = new ArrayList<>(options);


fridrich:
The https://github.com/apache/maven-javadoc-plugin/commit/d2dd5321dc02e793d87c722224c40d94a607cb51 brought an inconsistency that I saw with a project containing chinese characters and I made it consistent with https://github.com/apache/maven-javadoc-plugin/commit/33c9f01af9a3d4d28decbabd0bc02c4b3a875c2d

I assume that this bug comes from something related to the way we are determining the charset in that getDataCharset function. It might be needed to make the decision making depending not only on java version but also on os/arch.

I will test this fix. If it is not making the previous bug come back, we can just leave it as it is, but if it regresses, we should look for a proper fix.

fridrich:
Actually, as I look at it more and more and as I understand it better, I think this particular fix is the right one.

gnodet:
> I mean, I run the ITs with openjdk 8 11 and 17 and they passed.

On Windows, where the default encoding is not UTF-8 ?

I think the problem is that javadoc does expect a certain encoding, so we have no choice but to actually use that one.  If we always write in UTF-8, I don't see how that would work.

fridrich:
> > I mean, I run the ITs with openjdk 8 11 and 17 and they passed.
> 
> On Windows ?

Indeed, no

fridrich:
Or, we extract the getDataCharset() somewhere in SystemUtils or so, make it  public and use it everywhere where we need to use the charset

PR #1278, opened 2025-10-17
Use consistently the encoding expected by Javadoc for reading and writing of data

fridrich:

- [x] Your pull request should address just one issue, without pulling in other changes.
  Note that commits might be squashed by a maintainer on merge.
- [ ] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [x] Run `mvn verify` to make sure basic checks pass.
  A more thorough check will be performed on your pull request automatically.
- [x] You have run the integration tests successfully (`mvn -Prun-its verify`).

- [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -4149,14 +4149,8 @@ public abstract class AbstractJavadocMojo extends AbstractMojo {
         StringBuilder options = new StringBuilder();
         options.append(StringUtils.join(arguments.iterator(), SystemUtils.LINE_SEPARATOR));
 
-        Charset outputFileEncoding;
-        if (JAVA_VERSION.isAtLeast("9") && JAVA_VERSION.isBefore("12")) {
-            outputFileEncoding = StandardCharsets.UTF_8;
-        } else {
-            outputFileEncoding = Charset.defaultCharset();
-        }
         try {
-            Files.write(optionsFile.toPath(), Collections.singleton(options), outputFileEncoding);
+            Files.write(optionsFile.toPath(), Collections.singleton(options), SystemUtils.getExpectedEncoding());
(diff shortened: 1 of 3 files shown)


fridrich:
See: https://github.com/apache/maven-javadoc-plugin/pull/1274

fridrich:
@gnodet what do you think about this one?

gnodet:
> @gnodet what do you think about this one?

I haven't looked if we cover all the needed calls to files being read or written, but this looks good at first glance.

fridrich:
what would be the best name for this method? the getDataCharset was nice for the StaleHelper, but more generic would be something like getSystemEncoding or even more accurate getEncodingToUse... I have no idea :/

fridrich:
> > @gnodet what do you think about this one?
> 
> I haven't looked if we cover all the needed calls to files being read or written, but this looks good at first glance.

I grepped UTF_8 and then the files that had it, I checked the different calls. If this one passes all the ITs on all platforms, I will have to beautify it and will go through all of the source files looking for usual suspects.

fridrich:
@gnodet there are two occurrences of StandardCharsets.UTF_8 in JavadocUtil.java, but they have nothing to do with the filesystem encoding. In both cases, it is the encoding of maven invokerLog used to convert the byte[] resulting from reading it by Files.readAllBytes into a String. They should be left as they are IMHO.

fridrich:
> @gnodet there are two occurrences of StandardCharsets.UTF_8 in JavadocUtil.java, but they have nothing to do with the filesystem encoding. In both cases, it is the encoding of maven invokerLog used to convert the byte[] resulting from reading it by Files.readAllBytes into a String. They should be left as they are IMHO.

And I was overkeen in one place in AbstractJavadocMojo.java where the situation is the same

slachiewicz (approved):

PR #1279, opened 2025-10-22
chore: remove junit 3 references

sparsick:


erik-meuwese-topicus:
what is the purpose of empty classes `AppTest` and methods without asserts like `testMJAVADOC180()`?

slawekjaranowski:
@sparsick milestones was fixed - we need one current in open state

PR #1313, opened 2026-04-01
docletPath with relative path resolves incorrectly causing AssertionE…

martinaldrin:
<docletPath> with a relative path is written verbatim to the javadoc options file. The javadoc tool runs from target/reports/testapidocs, so a relative path like target/test-classes resolves to 
target/reports/testapidocs/target/test-classes — which doesn't exist. This causes the wrong class to be loaded from the classpath, which on Java 25 triggers a fatal assertion error:

error: fatal error encountered: java.lang.AssertionError: ignored flag: -Xlint:-options

On Java 21 and earlier the wrong class was silently loaded without crashing.

Fix: In AbstractJavadocMojo.getDocletPath(), resolve docletPath entries through JavadocUtil.prunePaths() (which resolves relative paths against project.getBasedir()) instead of passing them verbatim via 
unifyPathSeparator(). This is consistent with how other path parameters are already handled.

Note: testJavadocResourcesWithExcludes fails on a second consecutive mvn verify without clean due to a pre-existing stale data/test isolation issue in the suite — unrelated to this fix. mvn clean verify passes.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

Checklist status:
- ✅ Addresses one issue only
- ✅ Unit test added: JavadocUtilTest.testPrunePathsResolvesRelativePaths() — fails without the fix, passes with it
- ✅ mvn clean verify passes
- ⚠️ Integration tests (mvn -Prun-its verify) — please indicate if you've run these
- ✅ ~10 lines of code change — no ICLA required
- ✅ Licensed under Apache License 2.0

### DIFF
--- a/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
+++ b/src/main/java/org/apache/maven/plugins/javadoc/AbstractJavadocMojo.java
@@ -2892,7 +2892,9 @@ public abstract class AbstractJavadocMojo extends AbstractMojo {
         }
 
         if (!(docletPath == null || docletPath.isEmpty())) {
-            pathParts.add(JavadocUtil.unifyPathSeparator(docletPath));
+            for (Path path : JavadocUtil.prunePaths(project, Arrays.asList(JavadocUtil.splitPath(docletPath)), true)) {
+                pathParts.add(path.toString());
+            }
         }
 
         String path = StringUtils.join(pathParts.iterator(), File.pathSeparator);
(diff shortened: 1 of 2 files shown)


martinaldrin:
@slachiewicz I guess the failure is not related to my commit, saw similar failure on other PRs as well

martinaldrin:
@slachiewicz is there any plan for a new release in coming weeks?

PR #1328, opened 2026-06-03
Support multi-module builds with multiple versions of the same dependency in `javadoc:aggregate`, avoiding compile errors

### LINKED ISSUE #1327: `javadoc:aggregate` fails with compile errors when multi-module project uses different version of the same dependency
### Affected version

3.12.0

### Bug description

Imagine a multi-module project where different modules use different versions of the same dependency - e.g. connectors to version `1` and `2` of some service.

The `javadoc:aggregate` prunes the classpath supplied to the `javadoc` tool to include only the _latest_ version of a dependency - which can lead to compile errors when referencing something exclusively in the older version.

JackPGreen:
The dependency tree was being pruned - in case of multiple versions of the same dependency, only the latest version was retained.

However, this is unnecessary and can cause compile errors - instead, all of the dependencies should be added to the classpath of `javadoc`.

Fixes https://github.com/apache/maven-javadoc-plugin/issues/1327


JackPGreen:
I've raised multiple (unrelated) PRs and all of the `verify` jobs have failed with the [same error](https://github.com/apache/maven-javadoc-plugin/actions/runs/26941766513/job/79980895978?pr=1328) in `MJAVADOC-427` integration test in the `ubuntu-latest jdk-17-zulu 3.9.16` environment - I suspect there's an issue with the test itself.

JackPGreen:
@slachiewicz are you able to help with the next steps of getting this merged, please?

PR #1350, opened 2026-08-04
Verify the build against Maven 4 as well

slachiewicz:
The plugin is currently only verified against Maven 3. Adding `maven4-enabled: true` makes the shared verify workflow put the Maven 4 version into the build matrix alongside `3.10.0-rc-1`, so regressions under Maven 4 show up here instead of downstream. The shared workflow adds the `jdk: 8` / Maven 4 exclusion by itself, so nothing else has to change in the matrix.

The trailing comma in `matrix-exclude` has to go at the same time. Without `maven4-enabled` that value is echoed verbatim, so the invalid JSON went unnoticed; with the flag on, the shared workflow pipes it through `jq` to append its own exclusion, and `jq` rejects it:

```
$ echo "$matrix_exclude" | jq -c '. + [{"jdk":"8","maven":"4.0.0-rc-6"}]'
jq: parse error: Expected another array element at line 7, column 7
```

Verified locally with Maven 4.0.0-rc-6 on JDK 21: `clean verify` succeeds and all unit tests pass.


Bukama:
Duplicate of https://github.com/apache/maven-javadoc-plugin/pull/1323 But no strong opinion on what to keep. 1323 has some analytics on why it fails with (before rc6)

